PR #1, opened 2015-07-22
[Us 24] Change flow for unregistered user

andrii-r:
throw WrongPathException when unknown path parameter is passed, add 'guest' parameter to UserBean and UserDetails in API response, add api method and functionality for creating temporary user, split create and delete methods in UserDao, avoid creation of new public game for guest users

ghfilter
- src/main/java/.*
- src/test/.*


README.md
```diff
@@ -9,6 +9,10 @@ List of methods available via http with real examples can be found under the fol
 http://localhost:8091/api
 ```
 
+## How to create branch
+
+Don't know
+
```
andrii-r: Add branch creation details

README.md
```diff
@@ -9,6 +9,10 @@ List of methods available via http with real examples can be found under the fol
 http://localhost:8091/api
 ```
 
+## How to create branch
```
ypetrenko: all is good)

README.md
andrii-r: test

andrii-r: test1

andrii-r: test3

src/main/java/catan/controllers/ExceptionHandlerController.java
```diff
@@ -13,6 +14,14 @@
 
 @ControllerAdvice
 public class ExceptionHandlerController {
+
+    @ExceptionHandler(WrongPathException.class)
+    public ResponseEntity handleWrongPathExceptions(WrongPathException e) {
+        HttpStatus status = HttpStatus.NOT_FOUND;
+
```
andrii-r: test1

andrii-r: asdasd

README.md
andrii-r: good

andrii-r:
ghfilter
- .*\.java
- lib/support/.*
- lib/myfile.js

andrii-r:
ghfilter
- .*\.java
- *.java

src/test/java/catan/controllers/user/UserControllerTest.java
```diff
@@ -141,4 +155,53 @@ public void shouldLogoutSuccessfullyWithHttpCode200WhenAnyTokenIsPassed() {
...
+    }
+
+    @Test
+    public void should_fail_when_register_and_login_guest_if_username_is_empty() {
+        registerAndLoginGuest("")
+                .then()
+                .statusCode(400)
+                .body("errorCode", equalTo("ERROR"));
+    }
+
```
andrii-r, line 207: I would also add a test case, when guest is registered, and after that trying to login guest as regular user, but maybe it is not needed

src/test/java/catan/controllers/FunctionalTestUtil.java
```diff
@@ -1,6 +1,9 @@
+// TODO: remove checking of statusCode in each method. (Needs to be approved)
+
 package catan.controllers;
```
andrii-r, line 1: not good practice to write comments at the begining of the file, better add it at the begining of class body

src/test/java/catan/controllers/user/UserControllerTest.java
```diff
@@ -8,20 +11,24 @@
...
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.isEmptyString;
+import static org.hamcrest.Matchers.*;
 
 
 @RunWith(SpringJUnit4ClassRunner.class)
 @SpringApplicationConfiguration(classes = ApplicationConfig.class)
 @WebIntegrationTest("server.port:8091")
-public class UserControllerTest {
+public class UserControllerTest extends FunctionalTestUtil {
```
andrii-r, line 18: if you are extending from this class, you can migrate all player details calls to this method to avoid code duplication

src/test/java/catan/controllers/user/UserControllerTest.java
```diff
@@ -141,4 +155,53 @@ public void shouldLogoutSuccessfullyWithHttpCode200WhenAnyTokenIsPassed() {
...
+                .then()
+                .statusCode(200)
+                .body("token", not(equalTo(isEmptyString())))
+                .extract()
+                .path("token");
+
+        getUserDetails(userToken)
+                .then()
+                .statusCode(200)
+                .body("guest", equalTo(true));
```
andrii-r, line 176: I would assert more details response fields. What if username is empty for guest user? this test will be passing, but it is wrong

src/test/java/catan/controllers/user/UserControllerTest.java
```diff
@@ -141,4 +155,53 @@ public void shouldLogoutSuccessfullyWithHttpCode200WhenAnyTokenIsPassed() {
...
+    // end of code which needs to be refactored (see the first line of the file)
+
+
+
+    @Test
+    public void should_successfully_register_and_login_guest() {
+        String userToken = registerAndLoginGuest(USER_NAME_GUEST_1)
+                .then()
+                .statusCode(200)
+                .body("token", not(equalTo(isEmptyString())))
```
andrii-r: Again:  Hamcrest comes with a library of useful matchers. Here are some of the most important ones.

Core
anything - always matches, useful if you don't care what the object under test is
describedAs - decorator to adding custom failure description
is - decorator to improve readability - see "Sugar", below
Logical
allOf - matches if all matchers match, short circuits (like Java &&)
anyOf - matches if any matchers match, short circuits (like Java ||)
not - matches if the wrapped matcher doesn't match and vice versa
Object
equalTo - test object equality using Object.equals
hasToString - test Object.toString
instanceOf, isCompatibleType - test type
notNullValue, nullValue - test for null
sameInstance - test object identity
Beans
hasProperty - test JavaBeans properties
Collections
array - test an array's elements against an array of matchers
hasEntry, hasKey, hasValue - test a map contains an entry, key or value
hasItem, hasItems - test a collection contains elements
hasItemInArray - test an array contains an element
Number
closeTo - test floating point values are close to a given value
greaterThan, greaterThanOrEqualTo, lessThan, lessThanOrEqualTo - test ordering
Text
equalToIgnoringCase - test string equality ignoring case
equalToIgnoringWhiteSpace - test string equality ignoring differences in runs of whitespace
containsString, endsWith, startsWith - test string matching

src/test/java/catan/controllers/game/CreateGameTest.java
```diff
@@ -35,26 +37,22 @@ public void setup() {
...
+    }
+
+    @Test
+    public void should_guest_successfully_create_new_private_game() {
+        String userToken = registerAndLoginGuest(USER_NAME_GUEST_1)
+                .path("token");
+
+        createNewGame(userToken, true)
+                .then()
+                .statusCode(200);
```
andrii-r, line 55: maybe assert gameId?

src/test/java/catan/controllers/user/UserControllerTest.java
skisor, line 207: I think user can login as regular user even he's been registered as guest before. Anyway he can log out and do all actions which are allowed for unauthorized user

src/test/java/catan/controllers/FunctionalTestUtil.java
skisor, line 1: didn't know that. will fix

skisor, line 1: What you think about the comment? Should the checking be removed? (some day)

src/test/java/catan/controllers/user/UserControllerTest.java
skisor, line 18: UserControllerTest includes only tests. I left TODO:

> // TODO: these tests need to be refactored: create methods like "register", "login", ... which will be returning "Response" objects. And use them.

src/test/java/catan/controllers/user/UserControllerTest.java
skisor, line 176: Should checking of all details be in every test? Username and other details are checked in other tests. The only difference is "guest=true". Are you sure? I'm not

src/test/java/catan/controllers/user/UserControllerTest.java
skisor: shto?

src/test/java/catan/controllers/game/CreateGameTest.java
skisor, line 55: why? it's already checked in the previous test

src/test/java/catan/controllers/user/UserControllerTest.java
andrii-r, line 207: how can he login as a regular user? first of all we don;t know it as there is no such test case.
the seccond, if we allow user login for guests as general user, what password should he input? if it is empty, anyone will be able to login with his username))

src/test/java/catan/controllers/FunctionalTestUtil.java
andrii-r, line 1: I think we can remove it if we have assertion of object returned. In case of 'void' method, when there is no body, and nothing to assert, we should assert statusCode to make sure that is no errors and server replied with OK status

src/test/java/catan/controllers/user/UserControllerTest.java
andrii-r, line 18: ah, didn't see anything related to 'details' method in your TODO comments, thats why added this comment, but if you meant that everythig should be replaced, than ok)

src/test/java/catan/controllers/user/UserControllerTest.java
skisor, line 207: did you mix up login and registration in the first comment?

skisor, line 207: > if we allow user login for guests as general user, what password should he input?

I meant login by other username and password

src/test/java/catan/controllers/user/UserControllerTest.java
andrii-r, line 176: there may be different list of returned parameters for guest and user. but in this particular case, we can leave it as is

src/test/java/catan/controllers/user/UserControllerTest.java
andrii-r: this is a list of available hamcrest matchers that can be used instead of this shit: not(equalTo(isEmptyString()))

src/test/java/catan/controllers/game/CreateGameTest.java
andrii-r, line 55: actually you are right, we are asserting that in other test cases, but more assertions is better) we can leave it as is, this is just my purpose

src/test/java/catan/controllers/user/UserControllerTest.java
skisor: done

PR #5, opened 2015-08-18
Us 31 game dashboard and redirect

skisor:


src/main/webapp/features/screens/game/Game.controller.js
```diff
@@ -1,20 +1,27 @@
...
-                    $scope.gameDetails = response.data;
-                }, function(response) {
-                    alert('Getting Game Details Error: ' + ((response.data.errorCode) ? response.data.errorCode : 'unknown'));
-                    $state.go("start");
-                });
-        };
+        $scope.game = Game(+$stateParams.gameId);
 
-        $scope.updateGameDetails();
+        $scope.game.load()
```
andrii-r: I don't like that we call load() method of model to update this model, maybe we should have separate service, that load and updates model?

src/main/webapp/features/screens/game/Game.controller.js
```diff
@@ -1,20 +1,27 @@
...
+            .then(null, function() {
+                alert('Getting Game Details Error');
+                $state.go("start");
+            });
+
+        $scope.game.startUpdating(GAME_UPDATE_DELAY, null, function() {
```
andrii-r: the same here, maybe service should update the model, but not itself?

src/main/webapp/features/screens/game/dashboard/ct-game-dashboard.html
```diff
@@ -0,0 +1,3 @@
+game dashboard
+<br><br>
+object "game" is available here
```
andrii-r, line 3: it would be good to have some stub here, that represents game map overview, and details of the game in text format, jsut to be sure, that game is really avaliable here

src/main/webapp/features/screens/game/gameMapOverview/ct-game-map-overview.js
```diff
@@ -12,7 +12,7 @@ angular.module('catan')
                 },
                 link: function(scope, element) {
 
-                    var map = scope.game.map;
+                    var map = scope.game.get().map;
```
andrii-r: nice fix, I really like it!

src/main/webapp/features/screens/game/view.html
```diff
@@ -1,48 +1,3 @@
-<div class="row">
+<ct-game-room ng-if="game.isNew()" game="game"></ct-game-room>
```
andrii-r: I would rather split it on 2 different views resolved by ui-router on different pathes, for example 'game-details/123' and 'game-play/123'

src/main/webapp/features/shared/services/game/Game.service.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+angular.module('catan')
+        .factory('Game', ['GameModel', '$q', 'Remote', function (GameModel, $q, Remote) {
+
+            var Game = function(idOrDetails) {
```
andrii-r: wtf, what idOrDetails? it is very weird logic. as I understand, it is kind of constructor override, but it is better to make constructor take 3 parameters: id, details, status, so that we don't need strange names in code, like 'idOrDetails' and don't need this checks: typeof idOrDetails === "number" and typeof idOrDetails === "object"

src/main/webapp/features/shared/services/game/Game.service.js
```diff
@@ -0,0 +1,38 @@
...
+
+            Game.findPublic = function() {
+                return this.findByRemoteService("game", "listPublic");
+            };
+
+            Game.findCurrent = function() {
+                return this.findByRemoteService("game", "listCurrent");
+            };
+
+            Game.findByRemoteService = function(remoteGroup, remoteRequest) {
```
andrii-r: this can be simplified, to pass only scope.typeOfGames from ctGamesList directive, and in such case we can remove this logic: (scope.typeOfGames === 'CURRENT') ? 'findCurrent' : 'findPublic'
and remove 2 methods from this service: 'findPublic' and 'findCurrent' , so that we can leave only universal method that takes typeOfGames as a parameter.

src/main/webapp/features/shared/services/game/GameModel.service.js
```diff
@@ -0,0 +1,95 @@
+'use strict';
+
+angular.module('catan')
+        .factory('GameModel', ['$q', 'Remote', 'User', '$timeout', function ($q, Remote, User, $timeout) {
+
+            var GAME_PRIMARY_KEY = "gameId";
```
andrii-r, line 6: as I understand, it was introduced to a local variable so that it can be fixed quickly when we change id on server?

src/main/webapp/features/shared/services/game/GameModel.service.js
```diff
@@ -0,0 +1,95 @@
...
+                            deferred.resolve();
+                        }, function() {
+                            deferred.reject();
+                        });
+
+                        return deferred.promise;
+                    },
+                    startUpdating: function(delay, onEverySuccess, onEveryError) {
+                        var self = this;
+                        this.stopUpdating();
```
andrii-r: is there any case when we startUpdating more than one time?

src/main/webapp/features/shared/services/game/GameModel.service.js
```diff
@@ -0,0 +1,95 @@
...
+                        for (var i in details.gameUsers) {
+                            if (details.gameUsers[i].user.id === User.get().id) {
+                                return details.gameUsers[i];
+                            }
+                        }
+                        return false;
+                    },
+                    isCurrentUserReady: function() {
+                        var user = this.getCurrentUser();
+                        return (user) ? user.ready : undefined;
```
andrii-r: I very doubt that this method can return 'udefined', if method starts from 'is', it means that this method should return true or false

src/main/webapp/features/screens/game/room/ct-game-room.html
```diff
@@ -0,0 +1,45 @@
+<div class="row">
+
+    <div ng-if="game.isLoaded()">
```
andrii-r: why do we need this check? is there any cases when game is not loaded?

src/main/webapp/features/screens/game/Game.controller.js
skisor: needs to be discussed

src/main/webapp/features/screens/game/Game.controller.js
skisor: needs to be discussed

src/main/webapp/features/screens/game/dashboard/ct-game-dashboard.html
skisor, line 3: I swear it is... anyway that text will be removed in very next iteration

src/main/webapp/features/screens/game/view.html
skisor: why? It's not a game details, it's "game room". We can't see at the same time "game room" and "game playing"

src/main/webapp/features/shared/services/game/Game.service.js
skisor: but using becomes easier:
- var a = Game(123)    or   var b = Game({....})
- instead
- var a = Game(123, null)      or     var b = Game(null, {...})

and what is "status"?

src/main/webapp/features/shared/services/game/GameModel.service.js
skisor, line 6: seems so

src/main/webapp/features/shared/services/game/GameModel.service.js
skisor: don't understand the question

do you mean "case when we use function "startUpdating""?

src/main/webapp/features/shared/services/game/GameModel.service.js
skisor: but what if you can't determine?

src/main/webapp/features/screens/game/room/ct-game-room.html
skisor: probably when loading of the game is in progress

src/main/webapp/features/shared/services/game/Game.service.js
skisor: but then we will have to use it like this  [ct-game-list  type-of-games="listCurrent"]
it looks worse than [ct-game-list type-of-games="CURRENT"]

but I don't like it as well

go discuss that

src/main/webapp/features/screens/game/Game.controller.js
andrii-r: but lets discuss it before we merge this stuff to master

src/main/webapp/features/screens/game/dashboard/ct-game-dashboard.html
andrii-r, line 3: ok, lets leave it for next story

src/main/webapp/features/screens/game/view.html
andrii-r: I believe we can. Do you remember how tournaments happen in poker rooms? at the beginning you can see tournament details, and after game is started, you are redirected to your table, but you can still look at tournament details if you wish.

andrii-r: but we should think about strict flow to follow it and create appropriate story, so we can leave it as is.

src/main/webapp/features/shared/services/game/GameModel.service.js
andrii-r, line 6: can we use angular 'constant' initialization instead?

src/main/webapp/features/shared/services/game/GameModel.service.js
andrii-r: I mean, why we call this.stopUpdating() inside  startUpdating function? as I understand, we should do it only when startUpdating function was already called and we should cancel previously created timer. so the question is it really can happen, when we call startUpdating more than one time?

src/main/webapp/features/shared/services/game/GameModel.service.js
andrii-r: just return 'false', no? you never check that this method return 'undefined', you check only that it can be true or false, if method returns 'undefined', it will be converted to 'false' automatically, so I can't see any reason to return undefined, maybe because in Java we always return only true or false in methods that related to boolean logic.

src/main/webapp/features/screens/game/room/ct-game-room.html
andrii-r: to avoid empty fields when template loaded before we receive http response from server? In such case maybe we can show some progress bar or await on all pages, when are waiting for some http response from API calls? but we can leave it as is now

src/main/webapp/features/shared/services/game/Game.service.js
andrii-r: in the constructor you have main logic:

``` javascript
                if (typeof idOrDetails === "number") {
                    id = idOrDetails;
                } else if (typeof idOrDetails === "object") {
                    id = idOrDetails[GAME_PRIMARY_KEY];
                    details = idOrDetails;
                    status = STATUS_LOADED;
                }
```

So if you create constructor taking 3 parameters: (Id, details, status), you can remove this logic at all and just call constructors in 2 different ways: Game(123) or Game(123, {object,with,details}, "StatusValue") 
But this is just an assumption, how it can look like, but anyway, I don't like current implementation. idOrDetails - is not clear. You should change the logic in such way, that you don't need to create an Game only based on the id, without details. Maybe if we will have correct service that loads the game, we will remove this logic.

src/main/webapp/features/shared/services/game/Game.service.js
andrii-r: we can leave [ct-game-list type-of-games="CURRENT"], after that in ct-games-list.js, you should pass scope.typeOfGames directly in the method, that you have to create in Game service,for example Game.getListOfGames(scope.typeOfGames); so that you remove redundant logic in ct-games-list.js and redundant methods in Game.service.js, but lets discuss it in skype if my explanation is not clear

src/main/webapp/features/screens/game/Game.controller.js
andrii-r: decided to refactor model and service

src/main/webapp/features/screens/game/Game.controller.js
andrii-r: discussed

src/main/webapp/features/screens/game/view.html
andrii-r: lets leave as is as it helps to handle states much easier

src/main/webapp/features/shared/services/game/Game.service.js
andrii-r: solved

src/main/webapp/features/shared/services/game/Game.service.js
andrii-r: pass CURRENT value to Game.service directly

src/main/webapp/features/shared/services/game/GameModel.service.js
andrii-r: ok

src/main/webapp/features/shared/services/game/GameModel.service.js
andrii-r: return true or false

andrii-r: fix return value of getCurrentUser, return null if not found

PR #6, opened 2015-09-02
Us 27 add resources

ypetrenko:


src/main/java/catan/domain/model/game/GameUserBean.java
```diff
@@ -43,6 +42,9 @@
     @Column(name = "MOVE_ORDER", unique = false, nullable = false)
     private int moveOrder;
 
+    @ManyToOne(cascade = CascadeType.ALL)
```
andrii-r: Are you sure that we have to provide ManyToOne relationship here? I think each game user hase only one set of resources and each set of resources can belong only to one gameUser.

src/main/java/catan/domain/transfer/output/game/GameDetails.java
```diff
@@ -25,7 +25,7 @@
     public GameDetails() {
     }
 
-    public GameDetails(GameBean game) {
+    public GameDetails(GameBean game, int userId) {
```
andrii-r: rename this field to something like 'userWhoRequestedGameDetails' but shorter

src/main/java/catan/domain/transfer/output/game/GameUserDetails.java
```diff
@@ -8,15 +8,19 @@
     private int colorId;
     private boolean ready;
     private int moveOrder;
+    private ResourcesDetails resources;
 
     public GameUserDetails() {
     }
 
-    public GameUserDetails(GameUserBean userBean) {
+    public GameUserDetails(GameUserBean userBean, int userId) {
```
andrii-r: the same here, please rename parameter to understand in code, what does it mean and why we need to check it)

src/main/java/catan/services/impl/GameServiceImpl.java
```diff
@@ -293,8 +293,13 @@ private void startGame(GameBean game) {
                 return;
             }
         }
-
         log.debug("All players is ready");
+
+        log.debug("Setting resources to zero");
```
andrii-r: maybe it is better to set resources when gameuser is created to reduce steps during game creation? but it just point to discuss, we don;t need t change it right now

src/main/java/catan/controllers/GameController.java
```diff
@@ -46,12 +46,19 @@ public GameIdDetails createNewGame(@RequestParam(value = "token", required = fal
...
         switch (GameListType.fromValue(gameListType)) {
             case CURRENT:
-                UserBean user = authenticationService.authenticateUserByToken(token);
+                user = authenticationService.authenticateUserByToken(token);
+                userId = user.getId();
                 games = gameService.getListOfGamesWithJoinedUser(user);
                 break;
             case PUBLIC:
+                try {
+                    user = authenticationService.authenticateUserByToken(token);
```
andrii-r: don't need this code duplication, we can move this logic out of switch block

src/main/java/catan/controllers/GameController.java
```diff
@@ -46,12 +46,19 @@ public GameIdDetails createNewGame(@RequestParam(value = "token", required = fal
                                             @RequestParam(value = "token", required = false) String token) throws AuthenticationException, GameException, WrongPathException {
 
         List<GameBean> games;
+        UserBean user;
+        int userId = 0;
```
andrii-r: don't thinl that we need this variable, we can get value just via accessing the getter

src/main/java/catan/controllers/GameController.java
```diff
@@ -61,7 +68,7 @@ public GameIdDetails createNewGame(@RequestParam(value = "token", required = fal
 
         List<GameDetails> gamesToReturn = new ArrayList<GameDetails>();
         for (GameBean game : games) {
-            gamesToReturn.add(new GameDetails(game));
+            gamesToReturn.add(new GameDetails(game, userId));
```
andrii-r: btw, do we really need list of resources when retrieving list of games? I think we just need it when player tries to get details of some particular game

src/main/java/catan/controllers/GameController.java
```diff
@@ -95,8 +102,9 @@ public GameDetails getGameDetails(@RequestParam(value = "token", required = fals
                                       @RequestParam("gameId") String gameId) throws AuthenticationException, GameException {
         UserBean user = authenticationService.authenticateUserByToken(token);
         GameBean game = gameService.getGameByGameIdWithJoinedUser(user, gameId);
+        int userId = user.getId();
 
-        return new GameDetails(game);
+        return new GameDetails(game, userId);
```
andrii-r: I think this is the only one place where we need get list of resources

src/main/java/catan/controllers/GameController.java
```diff
@@ -95,8 +102,9 @@ public GameDetails getGameDetails(@RequestParam(value = "token", required = fals
                                       @RequestParam("gameId") String gameId) throws AuthenticationException, GameException {
         UserBean user = authenticationService.authenticateUserByToken(token);
         GameBean game = gameService.getGameByGameIdWithJoinedUser(user, gameId);
+        int userId = user.getId();
```
andrii-r: remove this variable and use 'user.getId()' directly, less lines, less code, and code becommes more readable. And rename 'user' to something more familiar with "user that initiated game details"

src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -241,11 +241,11 @@ public String toString() {
         return ToStringBuilder.reflectionToString(this, SHORT_PREFIX_STYLE);
     }
 
-    public List<GameUserDetails> getGameUserDetails() {
+    public List<GameUserDetails> getGameUserDetails(int userId) {
```
andrii-r: rename

src/main/java/catan/domain/model/game/GameUserBean.java
```diff
@@ -43,6 +42,9 @@
     @Column(name = "MOVE_ORDER", unique = false, nullable = false)
     private int moveOrder;
 
+    @Embedded
+    private ResourcesBean resources = new ResourcesBean();
```
andrii-r: we don't need to create new instance of ResourcesBean here since we create it in constructor

PR #8, opened 2015-09-12
Us 37 add development cards

andrii-r:


src/main/java/catan/domain/model/game/DevelopmentCardsBean.java
```diff
@@ -0,0 +1,83 @@
...
+    @Column(name = "VICTORY_POINT", nullable = false)
+    private int victoryPoint;
+
+    @Column(name = "ROAD_BUILDING", nullable = false)
+    private int roadBuilding;
+
+    @Column(name = "MONOPOLY", nullable = false)
+    private int monopoly;
+
+    @Column(name = "YEAR_OF_PLENTY", nullable = false)
```
andrii-r: Just thought that it is a good idea to name columns starting from some prefix, so that they will be placed close to each other in database, for example DEV_CARD_KNIGHT, DEV_CARD_VICTORY_POINT... 
The same approach should be applied to Resources

PR #9, opened 2015-09-12
Us 42 add build road functionality

andrii-r:


src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -31,7 +29,108 @@
...
+            }
+
+            if (node.getBuilding() == null) {
+                for (EdgeBean neighbourEdge : node.getEdges().all()) {
+                    if (neighbourEdge == null || neighbourEdge.equals(edgeToBuildOn)) {
+                        continue;
+                    }
+
+                    if (neighbourEdge.getBuilding() != null && neighbourEdge.getBuilding().getBuildingOwner().getUser().equals(user)) {
+                        nearNeighbourRoad = true;
```
ypetrenko, line 108: maybe you should break the loop here if you've got what you are intrested for?

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -31,7 +29,108 @@
...
+                continue;
+            }
+
+            if (node.getBuilding() == null) {
+                for (EdgeBean neighbourEdge : node.getEdges().all()) {
+                    if (neighbourEdge == null || neighbourEdge.equals(edgeToBuildOn)) {
+                        continue;
+                    }
+
+                    if (neighbourEdge.getBuilding() != null && neighbourEdge.getBuilding().getBuildingOwner().getUser().equals(user)) {
```
ypetrenko: you should also check if there no settlements that don't belong to user between neighbourEdge and edgeToBuildOn

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -31,7 +29,108 @@
...
+                        nearNeighbourRoad = true;
+                    }
+                }
+            } else {
+                if (node.getBuilding().getBuildingOwner().getUser().equals(user)) {
+                    nearNeighbourSettlement = true;
+                } else {
```
ypetrenko: should be discussed. i think that you can build your road close to opponents' settlement in case if you have contiguous road from another side of road that you want to build.

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -31,7 +29,108 @@
...
+                    log.debug("Cannot build road close to settles that don't belong to user");
+                    throw new PlayException(ERROR_CODE_ERROR);
+                }
+            }
+        }
+
+        if (!nearNeighbourRoad && !nearNeighbourSettlement) {
+            log.debug("Cannot build road close to settles that don't belong to user");
```
ypetrenko: please correct the log message

andrii-r: done

src/main/java/catan/services/impl/PlayServiceImpl.java
andrii-r, line 108: If I will break the loop, I will never get known about opposite node that can possibly have building that belongs to another player

ypetrenko, line 108: using break you would stop the cycle on line 102, but not 96.

andrii-r, line 108: sorry, missed it. in such case will fix it!

PR #10, opened 2015-09-12
Us 39 add functionality to end turn

andrii-r:


src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -31,7 +25,55 @@
...
+            log.debug("User is not joined to game with specified id %s", gameIdString);
+            throw new PlayException(ERROR_CODE_ERROR);
+        }
+
+        if (!game.getCurrentMove().equals(gameUserBean.getMoveOrder())) {
+            log.debug("It is not current turn of user {}", user.getUsername());
+            throw new PlayException(ERROR_CODE_ERROR);
+        }
 
+        //TODO: think about naming of method and namings of fields related to moveOrder and turn
```
ypetrenko, line 63: passTurn ?

andrii-r, line 63: pass means 'propustit', but we need here name functionality that can give Current Move To Next Player)

ypetrenko, line 63: pass usually means 'peredat', while miss is used for 'propustit'. But naming is not critical so never mind

andrii-r, line 63: but on of meanings is really 'propustit': http://c2n.me/3nFqiI3
so this name is not transparent ehough)  I think we should thinkg about it later, except if we have a great name)

PR #24, opened 2015-10-22
US-52_Add_build_actions_in_the_preparation_stage

andrii-r:


src/main/java/catan/services/PlayService.java
```diff
@@ -10,6 +10,8 @@
 
     void buildSettlement(UserBean user, String gameId, String nodeId) throws PlayException, GameException;
 
+    void buildCity(UserBean user, String gameIdString, String nodeIdString) throws PlayException, GameException;
```
andrii-r: I think we can remove 'String' in parameter name in the interface, it is clear that type of this argument has type String, but we should keep it in implementation of method to distinguish temp string values from real integer values

PR #29, opened 2015-11-03
Us 62 finish game when target victory points is achieved

andrii-r:


src/main/java/catan/domain/transfer/output/game/GameDetails.java
```diff
@@ -33,7 +33,7 @@ public GameDetails(GameBean game, int detailsRequesterId) {
         this.dateCreated = game.getDateCreated().getTime();
         this.dateStarted = game.getDateStarted() != null ? game.getDateStarted().getTime() : 0;
         this.status = game.getStatus().name();
-        this.gameUsers = game.getGameUserDetails(detailsRequesterId);
+        this.gameUsers = game.getGameUserDetails(detailsRequesterId, game.getStatus());
```
andrii-r: this is shit, we should not pass in the game's method parameter that we take from game

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -53,6 +53,8 @@ public void buildRoad(UserBean user, String gameIdString, String edgeIdString) t
         buildUtil.validateUserCanBuildRoanOnEdge(user, edgeToBuildOn);
         buildUtil.buildRoadOnEdge(user, edgeToBuildOn);
 
+        playUtil.finishGameIfTargetReached(game);
```
andrii-r: it is better to keep calls of playUtil methots behind each other, so ther code will look like:

```
    preparationStageUtil.updateCurrentCycleBuildingNumber(game);

    playUtil.finishGameIfTargetReached(game);
    playUtil.updateAvailableUserActions(game);
```

src/main/java/catan/services/util/play/MainStageUtil.java
```diff
@@ -32,7 +33,7 @@ public void updateNextMove(GameBean game) {
     public void updateAvailableUserActions(GameBean game) throws GameException {
 
         for (GameUserBean gameUser : game.getGameUsers()) {
-            if (gameUser.getMoveOrder() == game.getCurrentMove()) {
+            if (gameUser.getMoveOrder() == game.getCurrentMove() || !GameStatus.FINISHED.equals(game.getStatus())) {
```
andrii-r: It will work incorrectly!

src/main/java/catan/services/util/play/PlayUtil.java
```diff
@@ -52,6 +54,16 @@ public void updateAvailableUserActions(GameBean game) throws GameException {
         }
     }
 
+    public void finishGameIfTargetReached(GameBean game) {
```
andrii-r: rename method

src/main/java/catan/services/util/play/PlayUtil.java
```diff
@@ -52,6 +54,16 @@ public void updateAvailableUserActions(GameBean game) throws GameException {
...
+        for (GameUserBean gameUser : game.getGameUsers()) {
```
andrii-r: why do we iterate all player? only one plaer can reach victory points, and it is the player, that made action

PR #33, opened 2015-11-12
Us 70 buy development card

ypetrenko:


src/main/java/catan/controllers/PlayController.java
```diff
@@ -86,6 +87,16 @@ public void throwDice(@RequestParam(value = "token", required = false) String to
         playService.processAction(GameUserActionCode.THROW_DICE, user, gameId);
     }
 
+    @RequestMapping(value = "buy/card",
+            method = RequestMethod.POST,
+            produces = MediaType.APPLICATION_JSON_VALUE)
+    public Map<String, String> buyCard(@RequestParam(value = "token", required = false) String token,
```
andrii-r: I don't like that controller returns Map instead of DTO, we should create separate Java class for each controller's response

src/main/java/catan/domain/model/game/AvailableDevelopmentCards.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public int getYearOfPlenty() {
+        return yearOfPlenty;
+    }
+
+    public void setYearOfPlenty(int yearOfPlenty) {
+        this.yearOfPlenty = yearOfPlenty;
+    }
+
+    public Integer takeAvailableDevCardQuantity(DevelopmentCard developmentCard) {
```
andrii-r: I don't like this name:
1. it contains 'take'
2. it is too long
3. it contains name of class, that is redundand here

src/main/java/catan/domain/model/game/AvailableDevelopmentCards.java
```diff
@@ -0,0 +1,122 @@
...
+            case MONOPOLY:
+                return getMonopoly();
+            case YEAR_OF_PLENTY:
+                return getYearOfPlenty();
+            default:
+                return null;
+        }
+    }
+
+    public void decreaseAvailableDevCardQuantityByOne(DevelopmentCard developmentCard) {
```
andrii-r: I don't like this name. 
Btw, this is good example where word 'take' would match to explain what happens, for example: takeOneCardFromPlayer() - decreases his dev cards

But it was just an explanation, why I don't like 'take' in other methods, it doesn't mean that we should name this particular method in such way

src/main/java/catan/domain/model/game/UsersDevelopmentCards.java
```diff
@@ -0,0 +1,131 @@
...
+import catan.domain.model.game.types.DevelopmentCard;
+import org.apache.commons.lang.builder.ToStringBuilder;
+
+import javax.persistence.Column;
+import javax.persistence.Embeddable;
+
+import static org.apache.commons.lang.builder.ToStringStyle.SHORT_PREFIX_STYLE;
+
+@Embeddable
+public class UsersDevelopmentCards {
```
andrii-r: We can use the same class as for available development cards, we don't need 2 identical classes

src/main/java/catan/domain/model/game/UsersDevelopmentCards.java
```diff
@@ -0,0 +1,131 @@
...
+
+    public int getYearOfPlenty() {
+        return yearOfPlenty;
+    }
+
+    public void setYearOfPlenty(int yearOfPlenty) {
+        this.yearOfPlenty = yearOfPlenty;
+    }
+
+    public Integer takeUsersDevCardQuantity(DevelopmentCard developmentCard) {
```
andrii-r: take :(

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -125,12 +138,12 @@ private void endTurn(GameBean game) throws GameException {
             Integer newPreparationCycle = game.getPreparationCycle();
             boolean preparationFinished = newPreparationCycle == null && !game.getCurrentMove().equals(1);
             shouldUpdateNextMove = (preparationFinished || previousPreparationCycle.equals(newPreparationCycle));
+        } else {
+            game.setDiceThrown(false);
```
andrii-r: have we change it in scope of this story?

src/main/java/catan/domain/model/game/UsersDevelopmentCards.java
```diff
@@ -0,0 +1,131 @@
...
+            case MONOPOLY:
+                return getMonopoly();
+            case YEAR_OF_PLENTY:
+                return getYearOfPlenty();
+            default:
+                return null;
+        }
+    }
+
+    public void decreaseUsersDevCardQuantityByOne(DevelopmentCard developmentCard) {
```
andrii-r: do we need this method?

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -138,6 +151,43 @@ private void throwDice(GameBean game) {
         game.setDiceThrown(true);
     }
 
+    private void buyCard(UserBean user, GameBean game, Map<String, String> returnedParams) throws PlayException, GameException {
```
andrii-r: You do many actions here, that are not common for this service, I think we need something like CardUtil

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -138,6 +151,43 @@ private void throwDice(GameBean game) {
...
+    private void validateThereAreAvailableCards(int availableDevelopmentCardsQuantity) throws PlayException {
+        if (availableDevelopmentCardsQuantity == 0) {
+            log.debug("No available cards");
+            throw new PlayException(CARDS_ARE_OVER_ERROR);
+        }
+    }
+
+    private void populateAvailableDevCardsList(List<DevelopmentCard> availableDevelopmentCardsList, AvailableDevelopmentCards availableDevelopmentCards) {
+        for (DevelopmentCard developmentCard : DevelopmentCard.values()) {
+            for (int i = availableDevelopmentCards.takeAvailableDevCardQuantity(developmentCard); i > 0; i--) {
```
andrii-r: why do you have inverted iteration description here?

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -36,21 +41,24 @@
     private Logger log = LoggerFactory.getLogger(PlayService.class);
 
     public static final String ERROR_CODE_ERROR = "ERROR";
+    public static final String CARDS_ARE_OVER_ERROR = "CARDS_ARE_OVER";
```
andrii-r: was this name agreed? It seems very very very strange

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -138,6 +151,43 @@ private void throwDice(GameBean game) {
...
+        usersDevelopmentCards.increaseUsersDevCardQuantityByOne(obtainedDevelopmentCard);
+        availableDevelopmentCards.decreaseAvailableDevCardQuantityByOne(obtainedDevelopmentCard);
+
+        returnedParams.put("card", obtainedDevelopmentCard.name());
+    }
+
+    private DevelopmentCard pullDevelopmentCard(AvailableDevelopmentCards availableDevelopmentCards) throws PlayException {
+        List<DevelopmentCard> availableDevelopmentCardsList = new ArrayList<DevelopmentCard>();
+
+        populateAvailableDevCardsList(availableDevelopmentCardsList, availableDevelopmentCards);
```
andrii-r: it is better to return list in method rather than change existig one

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -138,6 +151,43 @@ private void throwDice(GameBean game) {
         game.setDiceThrown(true);
     }
 
+    private void buyCard(UserBean user, GameBean game, Map<String, String> returnedParams) throws PlayException, GameException {
+        AvailableDevelopmentCards availableDevelopmentCards = game.getAvailableDevelopmentCards();
+
+        DevelopmentCard obtainedDevelopmentCard = pullDevelopmentCard(availableDevelopmentCards);
```
andrii-r: I would think about better name of 'pullDevelopmentCard' method

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -138,6 +151,43 @@ private void throwDice(GameBean game) {
...
+
+        GameUserBean gameUserBean = gameUtil.getGameUserJoinedToGame(user, game);
```
andrii-r: you have fetched GameUserBean already. we don't have to do it again here

src/main/java/catan/domain/model/game/UsersDevelopmentCards.java
```diff
@@ -0,0 +1,131 @@
...
+        int newQuantityOfDevelopmentCard = takeUsersDevCardQuantity(developmentCard) - 1;
+        updateUsersDevCardQuantity(developmentCard, newQuantityOfDevelopmentCard);
+    }
+
+    public void increaseUsersDevCardQuantityByOne(DevelopmentCard developmentCard) {
```
andrii-r: i don't like name of this method

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: are there any difference?

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: I have changed it in the same way as in early 67 story, now it was changed. while merging the implementation from 67 would be inserted

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: that's why we have more utils then services:)

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: introducing new error code was agreed with front-end developer in common chat. he said: introduce new error code like "ZAKONCHILIS" and change story description.

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: i have no ideas how to name it better then 'pullDevelopmentCard'

src/main/java/catan/domain/transfer/output/game/BuyCardDetails.java
```diff
@@ -0,0 +1,20 @@
+package catan.domain.transfer.output.game;
+
+public class BuyCardDetails {
```
andrii-r: what about name BoughtCardDetails

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -138,6 +149,17 @@ private void throwDice(GameBean game) {
         game.setDiceThrown(true);
     }
 
+    private void buyCard(GameUserBean gameUser, GameBean game, Map<String, String> returnedParams) throws PlayException, GameException {
```
andrii-r, line 172: you don't take any resourcees from player when buying dev card

ypetrenko, line 172: i dont argee that it should be implemented in this story.
1. we doesn't give any resources to player yet, why it should be charged?
2. there are nothing noticed about it in story description.
3. we already have 65th story with the same functionality, why not to implement it there?

src/main/java/catan/services/util/play/CardUtil.java
```diff
@@ -0,0 +1,52 @@
...
+
+@Component
+public class CardUtil {
+    private Logger log = LoggerFactory.getLogger(CardUtil.class);
+
+    public static final String CARDS_ARE_OVER_ERROR = "CARDS_ARE_OVER";
+
+    private RandomUtil randomUtil;
+
+    public DevelopmentCard pullDevelopmentCard(DevelopmentCards availableDevelopmentCards) throws PlayException {
```
andrii-r: i would rename this method to something like 'chooseDevelopmentCard', as you don't pull anything from the parameter you pass

src/main/java/catan/services/util/play/CardUtil.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    private void validateThereAreAvailableCards(int availableDevelopmentCardsQuantity) throws PlayException {
+        if (availableDevelopmentCardsQuantity == 0) {
+            log.debug("No available cards");
+            throw new PlayException(CARDS_ARE_OVER_ERROR);
+        }
+    }
+
+    private List<DevelopmentCard> formAvailableDevCardsList(DevelopmentCards availableDevelopmentCards) {
```
andrii-r: we a;ready have convention, when we converting values of different fields to single ArrayList, and we name them as: 
'listAvailableDevCards'

src/main/java/catan/services/util/play/CardUtil.java
```diff
@@ -0,0 +1,52 @@
...
+    private RandomUtil randomUtil;
+
+    public DevelopmentCard pullDevelopmentCard(DevelopmentCards availableDevelopmentCards) throws PlayException {
+        List<DevelopmentCard> availableDevelopmentCardsList = formAvailableDevCardsList(availableDevelopmentCards);
+        validateThereAreAvailableCards(availableDevelopmentCardsList.size());
+
+        return randomUtil.pullRandomDevelopmentCard(availableDevelopmentCardsList);
+    }
+
+    private void validateThereAreAvailableCards(int availableDevelopmentCardsQuantity) throws PlayException {
```
andrii-r: since you create separate method, I would pass the whole list, and get size() of list directly here

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -138,6 +149,17 @@ private void throwDice(GameBean game) {
...
+        DevelopmentCards availableDevelopmentCards = game.getAvailableDevelopmentCards();
+        DevelopmentCard obtainedDevelopmentCard = cardUtil.pullDevelopmentCard(availableDevelopmentCards);
```
andrii-r: obtainedDevelopmentCard  - it is not obtained yet, it will be obtained after you give this resource to player

src/main/java/catan/services/impl/PlayServiceImpl.java
andrii-r, line 172: in such case we need a new story, 65 is already planned and taken in progress

ypetrenko, line 172: dont think so. we had a precedents earlier when we changed some details is stories in the air. BTW, actually story development did not started yet. Alex just assigned the task for himself one week ago but made nothing, he said about it. Branch even not created.

src/main/java/catan/services/util/play/CardUtil.java
ypetrenko: r u sure?) where can I find this convertion?

src/main/java/catan/services/impl/PlayServiceImpl.java
andrii-r, line 172: 1. when we changed some details, it was critical for particular story implementation. But now you are going to ADD (not "cahnge") new datails that have nothing to do with current story description. 
2. New details requires new planning, change of story scope, and creation of new tasks, such things should not be done during the sprint

src/test/java/catan/controllers/ctf/GameUserValidator.java
```diff
@@ -47,9 +47,9 @@ public Scenario doesntHaveAvailableAction(String action) {
         return scenario;
     }
 
-    public Scenario resourcesChanged(int brick, int wood, int sheep, int wheat, int stone) {
+    public Scenario resourcesQuantityChanged(int brick, int wood, int sheep, int wheat, int stone) {
```
ypetrenko: name of method doesn't say what specifically we should check here. it is not clear which parameters I should pass: new quantity of resources, or maybe the difference between old and new resources quantity

src/test/java/catan/controllers/ctf/GameUserValidator.java
```diff
@@ -61,4 +61,18 @@ public Scenario resourcesChanged(int brick, int wood, int sheep, int wheat, int
         return scenario;
     }
 
+    public Scenario devCardsQuantityChanged(int knight, int victoryPoint, int roadBuilding, int monopoly, int yearOfPlenty) {
```
ypetrenko: the same here.

src/test/java/catan/controllers/ctf/Scenario.java
```diff
@@ -199,51 +216,109 @@ public Scenario nextRandomDiceValues(List<Integer> nextRandomValues) {
...
     }
 
-    private void saveUsersResourcesValues(){
-        if(!trackResources){
+    private void saveUsersResourcesAndCardsValues() {
+        if (!trackResources && !trackCards) {
             return;
         }
 
         getGameDetails(1);
```
ypetrenko: what if we would have test cases with 2, 4, 5 or 6 players?
I think you shoud add iterator here instead of code doubling

src/test/java/catan/controllers/ctf/Scenario.java
```diff
@@ -199,51 +216,109 @@ public Scenario nextRandomDiceValues(List<Integer> nextRandomValues) {
...
-        int p1Brick = gameUser(1).getValueOf("resources.brick");
-        int p1Wood = gameUser(1).getValueOf("resources.wood");
-        int p1Sheep = gameUser(1).getValueOf("resources.sheep");
-        int p1Wheat = gameUser(1).getValueOf("resources.wheat");
-        int p1Stone = gameUser(1).getValueOf("resources.stone");
+        if (trackResources) {
+            int p1Brick = gameUser(1).getValueOf("resources.brick");
```
ypetrenko: i dont like that you use 'p' in the name of variables. is it the shortening of 'player'? in other places we use word 'user' (e.g. 244-248, 258-262 lines). I think that all namings should be in one style: either 'user' or 'player'

src/test/java/catan/controllers/testcases/play/BuyDevCardTest.java
```diff
@@ -0,0 +1,360 @@
...
+import catan.services.util.random.RandomUtilMock;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.SpringApplicationConfiguration;
+import org.springframework.boot.test.WebIntegrationTest;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import static catan.domain.model.dashboard.types.HexType.BRICK;
```
ypetrenko, line 16: types.HexType.*

src/test/java/catan/controllers/testcases/play/BuyDevCardTest.java
```diff
@@ -0,0 +1,360 @@
...
+import static catan.domain.model.dashboard.types.HexType.EMPTY;
+import static catan.domain.model.dashboard.types.HexType.SHEEP;
+import static catan.domain.model.dashboard.types.HexType.STONE;
+import static catan.domain.model.dashboard.types.HexType.WHEAT;
+import static catan.domain.model.dashboard.types.HexType.WOOD;
+import static catan.domain.model.game.types.DevelopmentCard.KNIGHT;
```
ypetrenko, line 22: types.DevelopmentCard.*

src/test/java/catan/controllers/testcases/play/BuyDevCardTest.java
```diff
@@ -0,0 +1,360 @@
...
+
+                .BUY_CARD(1).boughtCardIs(KNIGHT)
+                .BUY_CARD(1).boughtCardIs(VICTORY_POINT)
+                .BUY_CARD(1).boughtCardIs(ROAD_BUILDING)
+                .BUY_CARD(1).boughtCardIs(MONOPOLY)
+                .BUY_CARD(1).boughtCardIs(YEAR_OF_PLENTY);
+    }
+
+    @Test
+    public void should_increase_development_cards_when_user_buys_a_new_one() {
```
ypetrenko, line 115: it is the same test case as 'should_return_bought_card_in_response'. you should just unit the verifications

src/test/java/catan/controllers/testcases/play/BuyDevCardTest.java
```diff
@@ -0,0 +1,360 @@
...
+                .BUY_CARD(1)
+                .BUY_CARD(1)
+                .BUY_CARD(1)
+                .BUY_CARD(1)
+                .BUY_CARD(1)
+                .BUY_CARD(1)
+                .BUY_CARD(1)
+                .BUY_CARD(1)
+                .BUY_CARD(1)
+                .BUY_CARD(1).successfully()
```
ypetrenko, line 173: i think it would be good if you check the quantity of user's development cards here. it would represent that quantity of available development cards was correct.

src/test/java/catan/controllers/testcases/play/BuyDevCardTest.java
```diff
@@ -0,0 +1,360 @@
...
+
+                .BUY_CARD(1).failsWithError("CARDS_ARE_OVER");
+    }
+
+    //TODO: use this method when user will require resources for buying cards
+    private Scenario playPreparationStageAndBuildCity() {
```
ypetrenko, line 179: will you rework this method when user will require resources? qoz it looks very strange by now

src/test/java/catan/controllers/ctf/GameUserValidator.java
andrii-r: is this better:
resourcesQuantityChangedOn(int brickDiff, int woodDiff, int sheepDiff, int wheatDiff, int stoneDiff)

src/test/java/catan/controllers/ctf/GameUserValidator.java
andrii-r: will do in the same way as for resourcesQuantityChanged

src/test/java/catan/controllers/ctf/Scenario.java
andrii-r: good point, will change it

src/test/java/catan/controllers/ctf/Scenario.java
andrii-r: it was Alex who created those names, but I will change it if you like

src/test/java/catan/controllers/testcases/play/BuyDevCardTest.java
andrii-r, line 16: I never add or remove imports manually, IDEA handles it instead of me

src/test/java/catan/controllers/testcases/play/BuyDevCardTest.java
andrii-r, line 115: it is not the same test case, if I don't return bought card (like it happened today), this test case will pass and 'should_return_bought_card_in_response' will fail. And wise versa, if I return bought card but don't increase user's dev cards, current test case will fail and 'should_return_bought_card_in_response' will pass. 
So I don't think that we should combine test cases that test different things. But we can check everything in a single test case like it was done in PreparationStageTest, but I think that it was very bad idea

src/test/java/catan/controllers/testcases/play/BuyDevCardTest.java
andrii-r, line 173: There is no need to check constant value. Number of development cards doesn't affect frontend functionality, in this particular test case I should check behavior of system in case when cards are finished. I don't need to know the combination of dev cards: 14 5 2 2 2 or 25 0 0 0 0 - doesn't matter, I just need to know what error will be in response, as this affect front-end and shows the behavior of our system.

src/test/java/catan/controllers/testcases/play/BuyDevCardTest.java
andrii-r, line 179: yes, it was just a copy-paste

src/test/java/catan/controllers/ctf/GameUserValidator.java
ypetrenko: resourcesQuantityChangedBy sounds better, how do you think?

src/test/java/catan/controllers/ctf/Scenario.java
ypetrenko: yep

src/test/java/catan/controllers/testcases/play/BuyDevCardTest.java
ypetrenko, line 115: ok, lets keep it as is

src/test/java/catan/controllers/testcases/play/BuyDevCardTest.java
ypetrenko, line 173: ok

src/test/java/catan/controllers/ctf/GameUserValidator.java
andrii-r: good

PR #34, opened 2015-11-13
Us 67 implement throw dice action

andrii-r:


src/main/java/catan/domain/transfer/output/game/DiceDetails.java
```diff
@@ -0,0 +1,41 @@
...
+public class DiceDetails {
+    private Integer value;
+    private Integer firstValue;
+    private Integer secondValue;
+
+    public DiceDetails() {
+    }
+
+    public DiceDetails(Integer firstValue, Integer secondValue) {
+        this.value = (firstValue != null ? firstValue + secondValue : null);
```
andrii-r: I would avoid logic of "sum value" calculation in different places, why we can't just pass value as constructor parameter?

src/main/java/catan/domain/transfer/output/game/GameDetails.java
```diff
@@ -38,6 +39,7 @@ public GameDetails(GameBean game, int detailsRequesterId) {
         this.maxPlayers = game.getMaxPlayers();
         this.targetVictoryPoints = game.getTargetVictoryPoints();
         this.currentMove = game.getCurrentMove();
+        this.dice = new DiceDetails(game.getDiceFirstValue(), game.getDiceSecondValue());
```
andrii-r: do we need to put dice details when game is not started yet? it is just a question for discussion, but we have take a decision and follow it for other JSON elements that are not populated yet.

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -125,17 +131,26 @@ private void endTurn(GameBean game) throws GameException {
             Integer newPreparationCycle = game.getPreparationCycle();
             boolean preparationFinished = newPreparationCycle == null && !game.getCurrentMove().equals(1);
             shouldUpdateNextMove = (preparationFinished || previousPreparationCycle.equals(newPreparationCycle));
+        } else {
```
andrii-r: by adding "else" condition, logic becomes wired a bit, your implementation requires performing significant actions in different places of "endTurn" method. I would like to fully refector this method.

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -125,17 +131,26 @@ private void endTurn(GameBean game) throws GameException {
...
         if (shouldUpdateNextMove) {
             playUtil.updateNextMove(game);
-            //TODO: set it only in MAIN stage
-            game.setDiceThrown(false);
         }
     }
 
     private void throwDice(GameBean game) {
+        Integer diceFirstValue = randomUtil.pullRandomDiceNumber(new ArrayList<Integer>(asList(1, 2, 3, 4, 5, 6)));
+        Integer diceSecondValue = randomUtil.pullRandomDiceNumber(new ArrayList<Integer>(asList(1, 2, 3, 4, 5, 6)));
```
andrii-r: this method is used for other purposes, and according to comment, contact of this method can be changed soon so that we can't use it.

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -125,17 +131,26 @@ private void endTurn(GameBean game) throws GameException {
...
+        game.setDiceFirstValue(diceFirstValue);
+        game.setDiceSecondValue(diceSecondValue);
         game.setDiceThrown(true);
+        if (diceFirstValue + diceSecondValue != 7) {
```
andrii-r: maybe move this check to separate method, as it is not clear from the code, why exactly you exclude dice value 7

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -125,17 +131,26 @@ private void endTurn(GameBean game) throws GameException {
...
+            mainStageUtil.distributeResources(game);
```
andrii-r: 'distribute' sound like it is something related to goods trading) maybe change this word to "propagate"? or maybe "produce"? or leave it as is...

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -38,10 +42,12 @@
     public static final String ERROR_CODE_ERROR = "ERROR";
 
     private GameDao gameDao;
+    private RandomUtil randomUtil;
     private GameUtil gameUtil;
     private PlayUtil playUtil;
     private BuildUtil buildUtil;
     private PreparationStageUtil preparationStageUtil;
+    private MainStageUtil mainStageUtil;
```
andrii-r, line 50: since we inject preparationStageUtil, mainStageUtil and playUtil to playService and also preparationStageUtil and mainStageUtil to playUtil , I think we have wrong architecture, we should think how to refactor it

src/main/java/catan/services/util/play/MainStageUtil.java
```diff
@@ -30,6 +36,45 @@ public void updateNextMove(GameBean game) {
...
     }
 
+    public void distributeResources(GameBean game) {
+        Integer diceValue = game.getDiceFirstValue() != null ? game.getDiceFirstValue() + game.getDiceSecondValue() : 0;
+        for (HexBean hex : game.getHexes()) {
+            if (diceValue.equals(hex.getDice())) {
+                for (NodeBean node : hex.getNodes().listAllNotNullItems()) {
+                    if (node.getBuilding() != null) {
+                        giveResourceToPlayer(node.getBuilding().getBuildingOwner(), node.getBuilding().getBuilt(), hex.getResourceType());
+                    }
```
andrii-r: I see you like to have deep nesting logic) there are 6 tabs between border and giveResourceToPlayer() method call, it should tell you that this code is not readable

src/main/java/catan/services/util/play/MainStageUtil.java
```diff
@@ -30,6 +36,45 @@ public void updateNextMove(GameBean game) {
         game.setCurrentMove(nextMoveNumber);
     }
 
+    public void distributeResources(GameBean game) {
+        Integer diceValue = game.getDiceFirstValue() != null ? game.getDiceFirstValue() + game.getDiceSecondValue() : 0;
```
andrii-r: this is second place where we calculate sum, so we should avoid it

andrii-r: why you decided to put 0 instead of null?

andrii-r: maybe better call diceSumValue ?

src/main/java/catan/services/util/play/MainStageUtil.java
```diff
@@ -30,6 +36,45 @@ public void updateNextMove(GameBean game) {
...
+                int currentSheepQuantity = gameUser.getResources().getSheep();
+                gameUser.getResources().setSheep(currentSheepQuantity + resourceQuantity);
+                break;
+            case STONE:
+                int currentStoneQuantity = gameUser.getResources().getStone();
+                gameUser.getResources().setStone(currentStoneQuantity + resourceQuantity);
+                break;
+            case WHEAT:
+                int currentWheatQuantity = gameUser.getResources().getWheat();
+                gameUser.getResources().setWheat(currentWheatQuantity + resourceQuantity);
```
andrii-r: gameUser.getResources() is called 10 times in this method, maybe simplify it?

src/main/java/catan/domain/transfer/output/game/GameDetails.java
ypetrenko: as temporary decision, I refactored it so that "dice = null" in case if dices not thrown yet. IMHO, in future we shoud create separated JSONs for new/playing status and preparation/main stage

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: don't see the ways how to refactor it fully now. how about switch-case instead of if-else? i think it would be a little bit more readable.

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: ok i will create the new one, but name of old method randomUtil.pullRandomDiceNumber should be changed coz it is not correct.
done

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: done

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: produce is better then propagate) done

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko, line 50: yep. TODO added

src/main/java/catan/services/util/play/MainStageUtil.java
ypetrenko: well, it is only two blocks "cycle-condition" inside each other. any ideas gow to simplify it? 
To take out one of them to separate method is not a good decision coz readability would be worse.

src/main/java/catan/services/util/play/MainStageUtil.java
ypetrenko: done

src/main/java/catan/domain/transfer/output/game/DiceDetails.java
ypetrenko: fixed

src/main/java/catan/services/util/play/MainStageUtil.java
ypetrenko: fixed

src/main/java/catan/services/impl/PlayServiceImpl.java
andrii-r: reason of my comment not in readability, if-else condition reads good. I don't like that you do different actions in different places. Before you had only 2 places where state of game changed, now you have 3 places, that are being called in different cases

ypetrenko: ok, how about this interpretation? http://prntscr.com/92bu56

src/main/java/catan/services/impl/PlayServiceImpl.java
andrii-r: how are you going to call it?

src/main/java/catan/services/impl/PlayServiceImpl.java
andrii-r, line 50: do you have any ideas now?

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: maybe "pullRandomDiceNumber" for dices and "pullRandomHexDiceNumber" for hexes?

src/main/java/catan/services/util/play/MainStageUtil.java
andrii-r: there can be even one cycle-condition, but 6 tabs is not a good code. If you don't have any ideas, leave it as is

src/main/java/catan/services/impl/PlayServiceImpl.java
andrii-r: ok, lets leave it, but put 'updateNextMove' to the bottom

src/main/java/catan/services/impl/PlayServiceImpl.java
andrii-r: why do you need to pull dice number?

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: done. also game.setDiceThrown(false);
        game.setDiceFirstValue(null);
        game.setDiceSecondValue(null);
moved to separate method resetDices()

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko: pull -> get
public int getRandomDiceNumber() {
        return (int) Math.ceil(rvg.randomValue() \* 6);
    }

src/main/java/catan/services/impl/PlayServiceImpl.java
ypetrenko, line 50: not yet

src/main/java/catan/services/util/play/MainStageUtil.java
ypetrenko: ok i would leave it as is and add a new todo

src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -298,6 +320,14 @@ public void setNodes(Set<NodeBean> nodes) {
         this.nodes = nodes;
     }
 
+    public Integer getDiceValue() {
+        if (this.diceFirstValue != null && this.diceSecondValue != null) {
```
andrii-r: we can avoid using "else" in this method. And make check in another way, to check if values are null, then return null, as it is exceptional case.

src/main/java/catan/domain/transfer/output/game/GameDetails.java
```diff
@@ -38,6 +39,7 @@ public GameDetails(GameBean game, int detailsRequesterId) {
         this.maxPlayers = game.getMaxPlayers();
         this.targetVictoryPoints = game.getTargetVictoryPoints();
         this.currentMove = game.getCurrentMove();
+        this.dice = game.getDiceFirstValue() == null ? null : new DiceDetails(game.getDiceValue(), game.getDiceFirstValue(), game.getDiceSecondValue());
```
andrii-r: Andrii:
do we need to put dice details when game is not started yet? it is just a question for discussion, but we have take a decision and follow it for other JSON elements that are not populated yet.

Yurii:
as temporary decision, I refactored it so that "dice = null" in case if dices not thrown yet. IMHO, in future we shoud create separated JSONs for new/playing status and preparation/main stage

My questions to this commit:
1. Have we already discussed and taken the final decision? This solution can make troubles on front-end if it expects not-null value of JSON element 'dice' even if dice was not thrown.
2. Do you want to disucss format of response in different cases? we have already implemented the core and front-end architecture doesn't expect any significant changes in response, so if you would like to change the format we shuld discuss it as soon as possible but not later.

src/main/java/catan/services/util/play/PlayUtil.java
```diff
@@ -40,6 +40,12 @@ public void updateNextMove(GameBean game) throws GameException {
         }
     }
 
+    public void resetDices (GameBean game) {
```
andrii-r: was it agreed that we have to reset them when user ends his turn? I think we don't handle it on front-end

src/main/java/catan/domain/transfer/output/game/GameDetails.java
ypetrenko: This changes (dice=null) was discussed with front-end developer before
comit. Moreover, it was his propose:)
And lets leave the format of Details as is, I have changed my opinion.

src/main/java/catan/services/util/play/PlayUtil.java
ypetrenko: it is expected naturally, coz when the move is changed, old throw dice results became an invalid.  The same as preparationCycle and currentCycleBuildingNumber values became invalid after Preparation stage changes to Main and its values are nulling.

andrii-r: There is a bug happened on front-end after you have added reser

src/main/java/catan/domain/transfer/output/game/GameDetails.java
andrii-r: Still discussing

src/main/java/catan/services/util/play/PlayUtil.java
ypetrenko: Wat? Which one?

src/main/java/catan/domain/model/dashboard/HexBean.java
```diff
@@ -131,6 +132,17 @@ public void setNodes(VerticalLinks<NodeBean> nodes) {
         this.nodes = nodes;
     }
 
+    public Set<NodeBean> getNodesWithBuildings() {
```
andrii-r: it is not a good practice to name methods like field accessor (starting from 'get' + 'NameOfField'), in beans that are mapped to database, as different libraries for db-mapping can think that 'getNodesWithBuildings' is an accessor for field with name 'nodesWithBuildings', and will throw an exception when such field is not found. We already have some methods with such naming that are used for DTO building, but they are also needs to be replaced to another util class

src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -290,6 +300,17 @@ public void setHexes(Set<HexBean> hexes) {
         this.hexes = hexes;
     }
 
+    public Set<HexBean> getHexesWithCurrentDiceValue() {
```
andrii-r: method starts from 'get'

src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -298,6 +319,13 @@ public void setNodes(Set<NodeBean> nodes) {
         this.nodes = nodes;
     }
 
+    public Integer getDiceValue() {
```
andrii-r: method starts from 'get'

src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -290,6 +300,17 @@ public void setHexes(Set<HexBean> hexes) {
         this.hexes = hexes;
     }
 
+    public Set<HexBean> getHexesWithCurrentDiceValue() {
+        Set<HexBean> hexesWithDiceNumber = new HashSet<HexBean>();
```
andrii-r: for small amount of elements better use List rather than Set

src/main/java/catan/domain/model/dashboard/HexBean.java
```diff
@@ -131,6 +132,17 @@ public void setNodes(VerticalLinks<NodeBean> nodes) {
...
+        Set<NodeBean> nodesWithBuildings = new HashSet<NodeBean>();
```
andrii-r: for small amount of elements better use List rather than Set

src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -290,6 +300,17 @@ public void setHexes(Set<HexBean> hexes) {
...
+        for (HexBean hex : this.hexes) {
+            if (getDiceValue() != null && getDiceValue().equals(hex.getDice())) {
```
andrii-r: I would invert this check, as according to game logic you have to search for hexesWithDiceNumber only when diceValue not null

src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -290,6 +300,17 @@ public void setHexes(Set<HexBean> hexes) {
         this.hexes = hexes;
     }
 
+    public Set<HexBean> getHexesWithCurrentDiceValue() {
+        Set<HexBean> hexesWithDiceNumber = new HashSet<HexBean>();
+        for (HexBean hex : this.hexes) {
```
andrii-r: I think that you iterate values once again, when you return them, so what is the purpose of this method? to remove one "if" check, you have added 10 more new lines of code

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -115,27 +119,40 @@ private void buildCity(UserBean user, GameBean game, String nodeId) throws PlayE
...
+
+            case PREPARATION:
+                Integer previousPreparationCycle = game.getPreparationCycle();
+                preparationStageUtil.updateGameStageToMain(game); //TODO: move it to the end of method calls
+                preparationStageUtil.updateCurrentCycleInitialBuildingNumber(game);
+                preparationStageUtil.updatePreparationCycle(game);
+                Integer newPreparationCycle = game.getPreparationCycle();
+                boolean preparationFinished = newPreparationCycle == null && !game.getCurrentMove().equals(1);
+                if (preparationFinished || previousPreparationCycle.equals(newPreparationCycle)) {
+                    playUtil.updateNextMove(game);
```
andrii-r: since in this case block we call only methods from preparationStageUtil, maybe we should avoid calling playUtil here?

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -115,27 +119,40 @@ private void buildCity(UserBean user, GameBean game, String nodeId) throws PlayE
...
+                }
+                break;
 
-        if (shouldUpdateNextMove) {
-            playUtil.updateNextMove(game);
-            //TODO: set it only in MAIN stage
-            game.setDiceThrown(false);
+            case MAIN:
+                playUtil.resetDices(game);
+                playUtil.updateNextMove(game);
```
andrii-r: if we will call only preparationStageUtil in previous block, maybe we should call mainStageUtil here?

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -115,27 +119,40 @@ private void buildCity(UserBean user, GameBean game, String nodeId) throws PlayE
...
+                    playUtil.updateNextMove(game);
+                }
+                break;
 
-        if (shouldUpdateNextMove) {
-            playUtil.updateNextMove(game);
-            //TODO: set it only in MAIN stage
-            game.setDiceThrown(false);
+            case MAIN:
+                playUtil.resetDices(game);
```
andrii-r: if this method is called only in MAIN stage, maybe we should place this method to mainStageUtil?

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -115,27 +119,40 @@ private void buildCity(UserBean user, GameBean game, String nodeId) throws PlayE
...
+        Integer diceSecondValue = randomUtil.getRandomDiceNumber();
+        game.setDiceFirstValue(diceFirstValue);
+        game.setDiceSecondValue(diceSecondValue);
         game.setDiceThrown(true);
+        if (!robberActivity(diceFirstValue, diceSecondValue)) {
+            mainStageUtil.produceResources(game);
+        }
+    }
+
+    private boolean robberActivity(Integer diceFirstValue, Integer diceSecondValue) {
```
andrii-r: names of all our methods in this service before were verbs, bot now it is noun, I don't like it

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -171,12 +188,18 @@ private void validateActionIsAllowedForUser(GameUserBean gameUser, GameUserActio
     }
 
 
+
```
andrii-r: we already have 2 lines, you decided to add third?))

src/main/java/catan/domain/model/game/Resources.java
```diff
@@ -75,6 +76,38 @@ public void setStone(int stone) {
         this.stone = stone;
     }
 
+    public int getResource(HexType resource) {
```
andrii-r: 'get' in name for not accessor

src/main/java/catan/domain/model/game/Resources.java
```diff
@@ -75,6 +76,38 @@ public void setStone(int stone) {
...
+            case WOOD:
+                return getWood();
+            case SHEEP:
+                return getSheep();
+            case STONE:
+                return getStone();
+            case WHEAT:
+                return getWheat();
+            default:
+                return 0;
```
andrii-r: if I pass some wrong HexType that is not listed in the switch-case, should I get 0? I think it is incorrect

src/main/java/catan/domain/model/game/Resources.java
```diff
@@ -75,6 +76,38 @@ public void setStone(int stone) {
...
+        }
+    }
+
+    public void setResource(HexType resource, int resourceQuantity) {
```
andrii-r: 'set' in not mutator method

src/main/java/catan/domain/transfer/output/game/GameDetails.java
```diff
@@ -2,11 +2,9 @@
 
 import catan.domain.model.game.GameBean;
 import catan.domain.transfer.output.dashboard.MapDetails;
-import com.fasterxml.jackson.annotation.JsonInclude;
 
 import java.util.List;
 
-@JsonInclude(JsonInclude.Include.NON_NULL)
```
andrii-r, line 9: don't remove it!

src/main/java/catan/services/util/play/MainStageUtil.java
```diff
@@ -30,6 +35,17 @@ public void updateNextMove(GameBean game) {
         game.setCurrentMove(nextMoveNumber);
     }
 
+    public void produceResources(GameBean game) {
+        for (HexBean hex : game.getHexesWithCurrentDiceValue()) {
+            for (NodeBean node : hex.getNodesWithBuildings()) {
+                Resources userResources = node.getBuilding().getBuildingOwner().getResources();
+                int currentResourceQuantity = userResources.getResource(hex.getResourceType());
+                int resourceQuantityToAdd = node.getBuilding().getBuilt().equals(NodeBuiltType.SETTLEMENT) ? 1 : 2;
```
andrii-r: what if put resourceQuantity as a hardcoded value to enum definition?

src/main/java/catan/services/util/play/MainStageUtil.java
```diff
@@ -30,6 +35,17 @@ public void updateNextMove(GameBean game) {
         game.setCurrentMove(nextMoveNumber);
     }
 
+    public void produceResources(GameBean game) {
+        for (HexBean hex : game.getHexesWithCurrentDiceValue()) {
+            for (NodeBean node : hex.getNodesWithBuildings()) {
+                Resources userResources = node.getBuilding().getBuildingOwner().getResources();
```
andrii-r: intorduce one more variable for 'node.getBuilding()' that is called more than once, as you use it in very long constructions

src/main/java/catan/services/util/play/MainStageUtil.java
```diff
@@ -30,6 +35,17 @@ public void updateNextMove(GameBean game) {
...
+                int currentResourceQuantity = userResources.getResource(hex.getResourceType());
```
andrii-r: 'hex.getResourceType()' can be also put to a separate variable

src/main/java/catan/services/util/play/PlayUtil.java
andrii-r: bug that we are discussing third day

src/main/java/catan/services/util/random/RandomUtil.java
```diff
@@ -39,12 +37,17 @@ public HexType pullRandomHexType(List<HexType> possibleHexTypes) {
...
         int randomDiceNumber = (int) (rvg.randomValue() * possibleDiceNumbers.size());
 
         return possibleDiceNumbers.remove(randomDiceNumber);
     }
 
+    public int getRandomDiceNumber() {
+        return (int) Math.ceil(rvg.randomValue() * 6);
+    }
+
+    @Autowired
```
andrii-r: why did you decided to add it?

src/main/java/catan/services/util/random/RandomUtil.java
```diff
@@ -39,12 +37,17 @@ public HexType pullRandomHexType(List<HexType> possibleHexTypes) {
...
     //TODO: add rule to avoid placing 6 and 8 dice numbers close to each other
-    public int pullRandomDiceNumber(List<Integer> possibleDiceNumbers) {
+    public int pullRandomHexDiceNumber(List<Integer> possibleDiceNumbers) {
         int randomDiceNumber = (int) (rvg.randomValue() * possibleDiceNumbers.size());
 
         return possibleDiceNumbers.remove(randomDiceNumber);
     }
 
+    public int getRandomDiceNumber() {
+        return (int) Math.ceil(rvg.randomValue() * 6);
```
andrii-r: what value is returned when 'rvg.randomValue()' returns 0?

src/test/java/catan/services/impl/PlayServiceImplTest.java
```diff
@@ -1173,6 +1182,57 @@ public void shouldFailWhenUserThrowDiceNotAtHisTurn() throws GameException, Play
         }
     }
 
+    @Test
+    public void shouldPassWhenPlayerThrowDice() throws GameException, PlayException {
```
andrii-r, line 1186: this test case fails for me

ypetrenko, line 1186: fixed

src/main/java/catan/services/util/play/PlayUtil.java
ypetrenko: fixed

src/main/java/catan/domain/model/dashboard/types/NodeBuiltType.java
```diff
@@ -1,5 +1,10 @@
 package catan.domain.model.dashboard.types;
 
 public enum NodeBuiltType{
-    SETTLEMENT, CITY
+    SETTLEMENT,
+    CITY;
+
+    public int resourceQuantityToAdd() {
+        return this == SETTLEMENT ? 1 : 2;
```
andrii-r: remove logic from method, use enum parameters instead

src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -298,6 +308,26 @@ public void setNodes(Set<NodeBean> nodes) {
...
+                if (calculateDiceValue().equals(hex.getDice())) {
+                    hexesWithDiceNumber.add(hex);
+                }
+            }
+        }
+
+        return hexesWithDiceNumber;
+    }
+
+    public Integer calculateDiceValue() {
```
andrii-r: maybe name method as 'calculateDiceSumValue' ?

src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -298,6 +308,26 @@ public void setNodes(Set<NodeBean> nodes) {
         this.nodes = nodes;
     }
 
+    public List<HexBean> fetchHexesWithCurrentDiceValue() {
+        List<HexBean> hexesWithDiceNumber = new ArrayList<HexBean>();
+        if (calculateDiceValue() != null) {
+            for (HexBean hex : this.hexes) {
+                if (calculateDiceValue().equals(hex.getDice())) {
```
andrii-r: possible NPE as calculateDiceValue() can return null

andrii-r: numerous calculations for each loop

src/main/java/catan/domain/model/game/Resources.java
```diff
@@ -75,6 +76,43 @@ public void setStone(int stone) {
         this.stone = stone;
     }
 
+    public Integer findResource(HexType resource) {
```
andrii-r: we count number of resources that player has, not resource itself, this name does not fit this case

src/main/java/catan/domain/model/game/Resources.java
```diff
@@ -75,6 +76,43 @@ public void setStone(int stone) {
...
+            case STONE:
+                return getStone();
+            case WHEAT:
+                return getWheat();
+            default:
+                return null;
+        }
+    }
+
+    public void updateResource(HexType resource, int resourceQuantity) {
```
andrii-r: the same for this name

src/main/java/catan/domain/transfer/output/game/DiceDetails.java
```diff
@@ -0,0 +1,56 @@
...
+    public DiceDetails(boolean thrown, Integer value, Integer first, Integer second) {
+        this.thrown = thrown;
+        if (thrown) {
+            this.value = value;
+            this.first = first;
+            this.second = second;
+        }
+    }
+
+    public boolean getThrown() {
```
andrii-r: for boolean accessors we use 'is' instead of 'get'

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -115,27 +119,38 @@ private void buildCity(UserBean user, GameBean game, String nodeId) throws PlayE
...
     }
 
     private void throwDice(GameBean game) {
+        Integer diceFirstValue = randomUtil.getRandomDiceNumber();
+        Integer diceSecondValue = randomUtil.getRandomDiceNumber();
+        game.setDiceFirstValue(diceFirstValue);
+        game.setDiceSecondValue(diceSecondValue);
         game.setDiceThrown(true);
+        if (!isRobbersActivity(diceFirstValue, diceSecondValue)) {
+            mainStageUtil.produceResourcesForUsersThatHaveBuildingsCloseToActiveHexes(game.fetchHexesWithCurrentDiceValue());
```
andrii-r: this name was just an example)

andrii-r: why not to introduce a new variable for HexesWithCurrentDiceValue?

src/main/java/catan/domain/model/dashboard/types/NodeBuiltType.java
```diff
@@ -1,5 +1,16 @@
...
+    SETTLEMENT(1),
+    CITY(2);
+
+    private final int resourceQuantityToAdd;
+
+    NodeBuiltType(int resourceQuantityToAdd) {
+        this.resourceQuantityToAdd = resourceQuantityToAdd;
+    }
+
+    public int defineResourceQuantityToAdd() {
```
andrii-r: you can use 'get' here

src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -298,6 +308,27 @@ public void setNodes(Set<NodeBean> nodes) {
         this.nodes = nodes;
     }
 
+    public List<HexBean> fetchHexesWithCurrentDiceValue() {
+        List<HexBean> hexesWithDiceNumber = new ArrayList<HexBean>();
+        Integer diceSumValue = calculateDiceSumValue();
+        if (diceSumValue != null) {
+            for (HexBean hex : this.hexes) {
+                if (diceSumValue.equals(hex.getDice())) {
+                    hexesWithDiceNumber.add(hex);
```
andrii-r: very deep method call) but lets leave it to avoid new refactoring, that can become worth

src/main/java/catan/services/util/random/RandomUtil.java
```diff
@@ -39,12 +37,17 @@ public HexType pullRandomHexType(List<HexType> possibleHexTypes) {
...
-    public int pullRandomDiceNumber(List<Integer> possibleDiceNumbers) {
+    public int pullRandomHexDiceNumber(List<Integer> possibleDiceNumbers) {
         int randomDiceNumber = (int) (rvg.randomValue() * possibleDiceNumbers.size());
 
         return possibleDiceNumbers.remove(randomDiceNumber);
     }
 
+    public int getRandomDiceNumber() {
+        int diceNumber = (int) Math.ceil(rvg.randomValue() * 6);
+        return diceNumber == 0 ? 1 : diceNumber;
```
andrii-r: in such case chance of value 1 is greater than others)
I would avoid using additional convertions, can we just increase generated value by 1?

src/main/java/catan/domain/model/game/Resources.java
```diff
@@ -75,6 +76,43 @@ public void setStone(int stone) {
         this.stone = stone;
     }
 
+    public Integer takeResourceQuantity(HexType resource) {
```
andrii-r: take is not the best verb)) it is better to leave this name wihout werbs, since it returns only value without any logic, so that you will call this method gameuser.getResources().quantityOf(WOOD); what do you think?

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
```diff
@@ -10,6 +14,8 @@ angular.module('catan')
                 },
                 link: function(scope, element) {
 
+                    var hexesHighlightTimeout = null;
```
andrii-r: why do you need it here?

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
```diff
@@ -31,7 +43,29 @@ angular.module('catan')
...
                 }
             };
 
+            //TODO: move that code to some helper?
+            function highlightHexes(element, map, dice, timeout) {
+                var hexesToHighlight = map.hexes.filter(function(hex) {
+                    return (dice === 7) ? hex.robbed : !hex.robbed && hex.dice === dice;
+                });
+
+                $timeout.cancel(timeout);
```
andrii-r: if there is no timer, then we have nothing to cancel, so why do we need to pass timeout in highlightHexes( ) every time?

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
```diff
@@ -31,7 +43,29 @@ angular.module('catan')
                         var edgeId = angular.element(event.target).closest(DrawMapService.EDGE_SELECTOR).attr('edge-id');
                         SelectMapObjectService.select('edge', edgeId);
                     });
+
                 }
             };
 
+            //TODO: move that code to some helper?
```
andrii-r, line 48: yes, good idea, create some dashboard-specific helper

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
```diff
@@ -31,7 +43,29 @@ angular.module('catan')
...
+            function highlightHexes(element, map, dice, timeout) {
+                var hexesToHighlight = map.hexes.filter(function(hex) {
+                    return (dice === 7) ? hex.robbed : !hex.robbed && hex.dice === dice;
```
andrii-r, line 51: if dice is 7, we should return false anyway, but you return value of hex.robbed - 'robbed' value will never return 'true' for hex with dice 7

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
```diff
@@ -31,7 +43,29 @@ angular.module('catan')
...
+                });
+
+                $timeout.cancel(timeout);
+                element.find(DrawMapService.HEX_SELECTOR).removeClass(HEXES_HIGHLIGHT_CLASS);
```
andrii-r: do we need remove it, since it is already removed?

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
```diff
@@ -31,7 +43,29 @@ angular.module('catan')
...
+
+                var elementsToHighlight = angular.element([]);
+                hexesToHighlight.forEach(function(hex) {
+                    elementsToHighlight = elementsToHighlight.add(DrawMapService.HEX_SELECTOR + "[hex-id="+hex.hexId+"]");
```
andrii-r: why do you need set value of elementsToHighlight again and again? when you call add() method. isn't it extends content of elementsToHighlight ?

src/main/webapp/features/screens/game/dashboard/resourcesPanel/ct-resources-panel.js
```diff
@@ -9,9 +9,9 @@ angular.module('catan')
                 },
                 templateUrl: "/features/screens/game/dashboard/resourcesPanel/ct-resources-panel.html",
                 link: function(scope) {
-                    var currentGameUser = scope.game.getCurrentUser();
-                    if (currentGameUser) {
-                        scope.resources = currentGameUser.resources;
```
andrii-r, line 14: what is a purpose of these changes?

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
skisor: Hexes are animated when dice value is changed. If some animation starts when previous one is not finished yet, we should stop the previous one. Animation is provided with javascript timeout. "hexesHighlightTimeout" - is a "promise" which is using for stoping of timeout before a new timeout (animation) needs to be started.

Why is it here? Because it is a private attribute of each instance of the directive.

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
skisor, line 51: dice - current gotten dice number here; hex - each hex on the map.  Why 'hex.robbed' will never return 'true'?

If dice != 7 we highlight hexes which have dice = gotten dice number (only if they are not robbed). If dice = 7, we highlight "robbers" (currently we highlight not exactly robbers but hexes which are robbed, but I think it doesn't matter for now). Isn't it right?

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
skisor: because there can be timer which is still working for previous animation

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
skisor: If there was timer which was working for previous animation and we stop it without removing of HEXES_HIGHLIGHT_CLASS  class, then animated hexes will be still animated, but they shouldn't.

Actually, we can remove all this complex thing with timers, but then animation of different dice actions can be happening in the same time

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
skisor: It doesn't, https://api.jquery.com/add/

src/main/webapp/features/screens/game/dashboard/resourcesPanel/ct-resources-panel.js
skisor, line 14: resources wasn't updated on the view when they are changed. It is simplest way that I found to make them updatable

PR #36, opened 2015-11-22
[US-38] Front-end: Add development cards button and panel

andrii-r:


src/main/webapp/features/screens/game/dashboard/cardsPanel/ct-cards-panel.js
```diff
@@ -0,0 +1,19 @@
...
+            return {
+                restrict: 'E',
+                scope: {
+                    game: "="
+                },
+                templateUrl: "/features/screens/game/dashboard/cardsPanel/ct-cards-panel.html",
+                link: function(scope) {
+                    scope.cards = function() {
+                        var currentGameUser = scope.game.getCurrentUser();
+                        return (currentGameUser) ? currentGameUser.developmentCards : {};
```
andrii-r, line 15: why do we need this check? should be there any case when currentGameUser is undefined?

src/main/webapp/features/screens/game/dashboard/cardsPanel/ct-cards-panel.html
```diff
@@ -0,0 +1,38 @@
+<ct-modal-window modal-window-id="CARDS_PANEL">
+
+    <div class="ct-modal-window-bg" ct-modal-window-close></div>
+    <div class="ct-modal-window-content content">
+
+        <button ng-disabled="cards().knight===0">
```
andrii-r, line 6: we should disable buttons, but show quantity value when it is not user's turn - now it is possible to click on the button if it is not user's turn

skisor, line 6: The buttons are dummies. When using of dev cards appears, new available actions will appear too, and the buttons will be enabled/disabled depend on them. For now, "disabling" using only for showing buttons' style

src/main/webapp/features/screens/game/dashboard/cardsPanel/ct-cards-panel.js
skisor, line 15: When a spectators are added to the game. If functionality of spectators is needed in the future

andrii-r, line 15: they will not have 'cards' button available at all, so why do we need this empty list?

src/main/webapp/features/screens/game/dashboard/cardsPanel/ct-cards-panel.html
andrii-r, line 6: you already have enabled/disabled functionality, but if you are going to implement third step: 'clickable', in such case leave it as is

PR #38, opened 2015-12-05
Us 74 use card monopoly

ypetrenko:


src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -211,37 +210,32 @@ private void buyCard(GameUserBean gameUser, GameBean game, Resources usersResour
...
+        Integer takenResourcesCount = cardUtil.takeResourceFromRivalsAndGiveItAwayToPlayer(gameUser, game, resource);
+        log.debug("Player got {} of {} resources", takenResourcesCount, resource);
+
+        returnedParams.put("resourcesCount", takenResourcesCount.toString());
+
+        cardUtil.takeDevelopmentCardFromPlayer(gameUser, DevelopmentCard.MONOPOLY);
+        game.setDevelopmentCardUsed(true);
     }
 
     private HexType validateResourceType(String resourceString) throws PlayException {
```
andrii-r: it is not clear that method returns something, rename it to "toValidResourceType":
'to..ResourceType' - says that it will convert input parameter to something and return it
'...Valid...' - says that it will validate input parameter

src/test/java/catan/controllers/testcases/play/UseCardMonopolyTest.java
```diff
@@ -0,0 +1,420 @@
...
+
+    @Autowired
+    private RandomUtil randomUtil;
+
+    private Scenario scenario;
+
+    @Before
+    public void setup() {
+        if (!initialized) {
+            registerUser(USER_NAME_1, USER_PASSWORD_1);
```
ypetrenko: u used an old methods. i'll fix it by myself

src/test/java/catan/controllers/testcases/play/UseCardMonopolyTest.java
```diff
@@ -0,0 +1,420 @@
...
+import static catan.domain.model.game.types.DevelopmentCard.MONOPOLY;
+import static java.util.Arrays.asList;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+
+//@SpringApplicationConfiguration(classes = {TestApplicationConfig.class, RequestResponseLogger.class})  // if needed initial request and JSON response logging:
+//@SpringApplicationConfiguration(classes = TestApplicationConfig.class)
+@SpringApplicationConfiguration(classes = TestApplicationConfig.class)
+@WebIntegrationTest("server.port:8091")
+public class UseCardMonopolyTest extends PlayTestUtil {
```
ypetrenko, line 31: test case is missing:
should_successfully_get_quantity_of_taken_resources()
i'll add it

PR #39, opened 2015-12-06
Us 73 use card road building

ypetrenko:


src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -102,6 +101,9 @@
     @Column(name = "DEV_CARD_USED", unique = false, nullable = true)
     private Boolean developmentCardUsed;
 
+    @Column(name = "ROADS_TO_BUILD_MANDATORY", unique = false, nullable = true)
```
andrii-r: i would rename it to: mandatory roads to build
and why is this column belongs to GAME table? should it be part od GAME_USER?

src/main/java/catan/services/util/play/CardUtil.java
```diff
@@ -69,6 +71,24 @@ public Integer takeResourceFromRivalsAndGiveItAwayToPlayer(GameUserBean currentG
         return takenResourcesCount;
     }
 
+    public Integer defineRoadsQuantityToBuild(GameUserBean gameUser, GameBean game) throws PlayException {
+        List<EdgeBean> availableEdges = new ArrayList<EdgeBean>(game.fetchEdgesAccessibleForBuildingRoad(gameUser));
+        switch (availableEdges.size()) {
```
andrii-r: I would simplify this logic to use somehow pretty if-else statment rather than switch-case

PR #41, opened 2015-12-10
Us 77 move robber when dice is 7

andrii-r:


src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -200,7 +203,8 @@ private void throwDice(GameBean game) {
             List<HexBean> hexesWithCurrentDiceValue = game.fetchHexesWithCurrentDiceValue();
             log.debug("Hexes with current dice value:" + hexesWithCurrentDiceValue);
             mainStageUtil.produceResourcesFromActiveDiceHexes(hexesWithCurrentDiceValue);
-        } else{
+        } else {
```
andrii-r: I would invert this logic

src/main/webapp/features/screens/game/dashboard/gameMap/ct-game-map.js
```diff
@@ -42,6 +42,11 @@ angular.module('catan')
                         SelectService.select('edge', edgeId);
                     });
 
+                    element.on('click', DrawMapService.HEX_SELECTOR, function(event) {
```
andrii-r, line 45: what happens when I click on the robbed hex?

skisor, line 45: App will try to move robber to the same place and get error. In future, back-end should provide info about places which are available for moving robber here -- like in building of road/settlement/city.

skisor, line 45: I've created story for this

PR #43, opened 2015-12-16
Us 80 steal resource from player when move robber

andrii-r:


src/test/java/catan/controllers/testcases/play/MoveRobberTest.java
```diff
@@ -57,22 +59,200 @@ public void setup() {
     }
```
ypetrenko, line 58: how about such test cases as:
shouldSuccessfullyRobPlayerWhenHeIsOnlyOneRivalAtRobbedHexAndCurrentUserHasHisOwnBuildingAtRobbedHex
shouldFailWhenUserRobHimself
shouldFailWhenUserRobPlayerThatDoesNotHaveBuildingAtRobbedHex

PR #44, opened 2015-12-16
Us 76 kick off resources when dice is 7

ypetrenko:


src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -277,6 +280,87 @@ private void moveRobber(GameBean game, final String hexId) throws PlayException,
...
+    private void checkRobberShouldBeMovedMandatory(GameBean game) {
+        for (GameUserBean gameUserIterated : game.getGameUsers()) {
+            if (gameUserIterated.isKickingOffResourcesMandatory()) {
+                return;
+            }
+        }
+        game.setRobberShouldBeMovedMandatory(true);
+    }
+
+    private void validateSumOfResources(int sumOfUsersResources, int sumOfResourcesKickingOff) throws PlayException {
```
andrii-r: this name doesn't tell exactly what it checks

PR #46, opened 2016-01-05
Us 82 use knight card

andrii-r:


src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -274,6 +277,17 @@ private void useCardRoadBuilding(GameUserBean gameUser, GameBean game, Map<Strin
...
 
+    private void useCardKnight(GameUserBean gameUser, GameBean game) throws PlayException, GameException {
+        cardUtil.validateUserDidNotUsedCardsInCurrentTurn(game);
+        cardUtil.validateUserDidNotBoughtCardInCurrentTurn(gameUser, DevelopmentCard.KNIGHT);
+        cardUtil.takeDevelopmentCardFromPlayer(gameUser, DevelopmentCard.KNIGHT);
+
+        game.setRobberShouldBeMovedMandatory(true);
+        game.setDevelopmentCardUsed(true);
+
+        gameUser.getAchievements().setTotalUsedKnights(gameUser.getAchievements().getTotalUsedKnights() + 1);
```
ypetrenko: i would split this line into two for better readability. also gameUser.getAchievements() could be puted to variable

ypetrenko:
Are you going to add unit tests?

src/main/java/catan/services/impl/PlayServiceImpl.java
andrii-r: ok

PR #55, opened 2016-02-12
Us 92 calculate biggest army and consider it in victory points

ypetrenko:


src/main/java/catan/domain/model/game/GameBean.java
```diff
@@ -90,6 +90,9 @@
     @Column(name = "CURRENT_MOVE", unique = false, nullable = true)
     private Integer currentMove;
 
+    @Column(name = "OWNER_BIGGEST_ARMY", unique = false)
```
andrii-r: Not clear what exact identifier is used (color, gameuserId or gameId, or moveorder) . maybe specify it in field name, or just add a comment

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -306,6 +306,28 @@ private void useCardKnight(GameUserBean gameUser, GameBean game) throws PlayExce
         int totalUsedKnightsNew = achievements.getTotalUsedKnights() + 1;
 
         achievements.setTotalUsedKnights(totalUsedKnightsNew);
+        updateBiggestArmyOwner(gameUser, game);
+    }
+
+    private void updateBiggestArmyOwner(GameUserBean gameUser, GameBean game) {
```
andrii-r, line 312: I would reorganize logic in this method, as it is too wired

ypetrenko, line 312: done

PR #61, opened 2016-03-15
Us 94 display available building to choose when player robs somebody

andrii-r:


src/main/java/catan/services/util/play/ActionParamsUtil.java
```diff
@@ -64,6 +65,23 @@
         return edgeIdsToBuildOn;
     }
 
+    public List<Integer> calculateChoosePlayerToRobParams(GameUserBean gameUser) {
+        List<Integer> nodeIdsToBuildOn = new ArrayList<Integer>();
```
andrii-r: we don't build on this node) rename it accordingly

src/main/java/catan/services/util/play/ActionParamsUtil.java
```diff
@@ -64,6 +65,23 @@
...
+        for (HexBean hex : gameUser.getGame().getHexes()) {
+            if (!hex.isRobbed()) {
+                continue;
+            }
+            for (NodeBean node : hex.getNodes().listAllNotNullItems()) {
+                if (node.getBuilding() != null && !node.getBuilding().getBuildingOwner().equals(gameUser)) {
```
andrii-r, line 75: should we check for it if we call listAllNotNullItems()? 
`node.getBuilding() != null`

andrii-r, line 75: We have method hasBuildingBelongsToUser() instead of doing this check manually

PR #72, opened 2016-04-01
Us 83 execute immediately actions

skisor:


src/main/webapp/new/app/play/shared/services/action.service.ts
```diff
@@ -0,0 +1,91 @@
...
+                });
+        },
+        'END_TURN': (game: Game) => {
+            this._play.endTurn(game)
+                .then(() => this._gameService.refresh(game))
+                .catch(data => alert('End turn error: ' + ((data.errorCode) ? data.errorCode : 'unknown')));
+        }
+    };
+
+    run(code: string, param?: any) { //TODO: try to use ES6 spread instead "param?: any"
```
andrii-r: maybe name it as 'execute'?

skisor: changed

src/main/webapp/new/app/play/shared/services/action.service.ts
```diff
@@ -0,0 +1,91 @@
...
+                });
+        },
+        'END_TURN': (game: Game) => {
+            this._play.endTurn(game)
+                .then(() => this._gameService.refresh(game))
+                .catch(data => alert('End turn error: ' + ((data.errorCode) ? data.errorCode : 'unknown')));
+        }
+    };
+
+    execute(code: string, param?: any) { //TODO: try to use ES6 spread instead "param?: any"
```
andrii-r: you don't use this param enywhere

andrii-r: there is only Game as param, why not to specify this?

src/main/webapp/new/app/play/game-map/game-map.component.ts
```diff
@@ -83,9 +83,16 @@ export class GameMapComponent implements OnInit {
...
-        this._marking.onUpdate('nodes', () => this._drawMapMarkingService.updateByType(this._canvas, 'nodes', 'node'));
-        this._marking.onUpdate('edges', () => this._drawMapMarkingService.updateByType(this._canvas, 'edges', 'edge'));
+        let types = [
+            ['hexes', 'hex'],
+            ['nodes', 'node'],
+            ['edges', 'edge']
+        ];
+
+        types.forEach(type => {
+            this._drawMapMarkingService.updateByType(this._canvas, type[0], type[1]);
```
andrii-r, line 93: why do you need to update it in first time? would "onUpdate" subscription be enough?

skisor, line 93: If you update the page when you have mandatory AA (e.g BUILD_ROAD) then setting of marking happens before subscription on marking change (because game-map is created after parent 'play.component')

src/main/webapp/new/app/play/shared/services/action.service.ts
skisor: fixed

src/main/webapp/new/app/play/play.component.ts
```diff
@@ -40,23 +44,57 @@ import { DiceComponent } from './dice/dice.component';
...
 
-    constructor(private _templates: TemplatesService) { }
+    constructor(
+        private _notification: NotificationService,
+        private _authUser: AuthUserService,
+        private _action: ActionService,
+        private _templates: TemplatesService) { }
 
     ngOnInit() {
+        this.availableActions = this.game.getCurrentPlayer(this._authUser.get()).availableActions;
```
andrii-r: you use this variable below calling 'this._checkIfImmediateAndExecute();', move it down

PR #76, opened 2016-04-03
Us 116 roads around design

skisor:


src/main/webapp/new/app/shared/domain/game-map/edge.ts
```diff
@@ -80,6 +80,17 @@ export class Edge {
...
+    isJoint() {
+        let hexesCount = 0;
+        for (let i in this.hexes) {
+            hexesCount++;
+            if (hexesCount > 1) {
+                return true;
+            }
+        }
+        return false;
+    }
```
andrii-r, line 92: maybe we can replace this method with something like this: 
`return this.hexes.size > 0`

src/main/webapp/new/app/play/game-map/services/draw-map.service.ts
```diff
@@ -170,21 +170,75 @@ export class DrawMapService {
         this._dom.setAttribute(group, 'hex-id', <string>hex.id);
         this._dom.appendChild(canvas, group);
 
+        this._dom.setInnerHTML(group, this._getHexHTML(hex));
+
+        hex.onUpdate(() => this.updateHex(canvas, group, hex));
```
andrii-r: move all map elements onUpdate listeneres upper to the cycle

src/main/webapp/new/app/shared/domain/game-map/node.ts
```diff
@@ -149,6 +149,17 @@ export class Node {
...
 
+    isJoint() {
+        let hexesCount = 0;
+        for (let i in this.hexes) {
+            hexesCount++;
+            if (hexesCount > 1) {
+                return true;
+            }
+        }
+        return false;
```
andrii-r, line 160: maybe we can replace this method with something like this: 
`return this.hexes.size > 0`

src/main/webapp/new/app/shared/domain/game-map/edge.ts
skisor, line 92: but "hexes" is not an array, it's an object, which has 6 properties, which can be null or have a value. We can add method size() which calculate count, but it will look like same as "isJoint()"

skisor: will use Object.keys(obj).filter(key => obj[key]).length

src/main/webapp/new/app/play/game-map/services/draw-map.service.ts
skisor: done

PR #78, opened 2016-04-06
Us 110 show help message on the top when a mandatory action is run

ypetrenko:


src/main/java/catan/domain/model/user/UserBean.java
```diff
@@ -25,13 +27,21 @@
...
+
     public UserBean() {
     }
 
     public UserBean(String username, String password, boolean guest) {
         this.username = username;
         this.password = password;
         this.guest = guest;
+        this.language = "en";
+        this.country = "US";
```
andrii-r: why do you hardcode these values in constructor, I would move it somewhere higher, so that I can easily change it if needed

src/main/java/catan/domain/model/user/UserBean.java
```diff
@@ -66,6 +76,27 @@ public void setGuest(boolean guest) {
...
+
+    public String getCountry() {
+        return country;
+    }
+
+    public void setCountry(String country) {
+        this.country = country;
+    }
+
+    public ResourceBundle getMsgs() {
```
andrii-r: according to logic, it appears that GameUser is owner of the messages, but it is not true, I would remove this message, as it is just a a bean that is mapped to database table,  it is not service that can provide some messages

src/main/java/catan/services/impl/PlayServiceImpl.java
```diff
@@ -159,6 +161,12 @@ private void doAction(GameUserActionCode action, GameUserBean gameUser, GameBean
         }
     }
 
+    private void clearUsersMessages(GameBean game) {
```
andrii-r: move this method lover than method where you cal this particular one.

src/main/java/catan/services/util/play/PreparationStageUtil.java
```diff
@@ -118,25 +120,35 @@ public void updateAvailableActionsForAllUsers(GameBean game) throws GameExceptio
...
                         break;
                     case BUILD_ROAD:
                         ActionOnEdgeParams buildRoadParams = new ActionOnEdgeParams(actionParamsUtil.calculateBuildRoadParams(gameUser));
                         actionToAdd = new Action(actionCode, buildRoadParams);
+                        Object[] argsForBuildRoadPattern = {(getFirstActionCodeInCurrentCycle(game).equals("BUILD_CITY") ? 2 : 1), username};
+                        msgToShow = getMessagePattern(gameUser, "help_msg_build_road").format(argsForBuildRoadPattern);
                         break;
                     default:
                         actionToAdd = new Action(actionCode);
+                        msgToShow = null;
```
andrii-r: better initiate it in null value during variable definition

PR #85, opened 2016-04-20
Us 125 set limit on settlements count

ypetrenko:


src/main/java/catan/services/util/play/BuildUtil.java
```diff
@@ -172,4 +173,19 @@ private void validateNoBuildingsCloseToNode(NodeBean nodeToBuildOn) throws PlayE
             throw new PlayException(ERROR_CODE_ERROR);
         }
     }
+
+    public void validateSettlementLimitExceeded(GameUserBean gameUser) throws PlayException {
+        log.debug("Validating if player exceed build settlement count limit");
+
+        int userHasSettlements = gameUser.getBuildingsCount().getSettlements();
```
ypetrenko: name of variable sounds like it should be boolean type but not int

src/main/java/catan/services/util/play/ValidationUtil.java
```diff
@@ -9,6 +9,7 @@
 public class ValidationUtil {
 
     public static final String ERROR_CODE_ERROR = "ERROR";
+    public static final String SETTLEMENTS_LIMIT_IS_REACHED_ERROR = "SETTLEMENTS_LIMIT_IS_REACHED ";
```
ypetrenko: redundant space

